PR #8, opened 2020-06-09
Merging the initial repo with @rossbar's simple setup

### LINKED ISSUE #4: Add MyST integration
At the docs team meeting, we discussed merging the ideas from [this repo](https://github.com/rossbar/sample_notebook_tutorials_site) by @rossbar. This is an improvement in that it has a simple workflow for building and hopefully makes contributions easier.

melissawm:
I have included a few theme tweaks that I found make it more usable, but the theme choice is not final. Hoping this is a better solution. 

I kept the CS231 tutorial mostly so we can test the theme and workflow, but if it doesn't fit here we can fix it later.

Please take a look @rgommers @rossbar @bjnath :)

Closes #4


rossbar:
LGTM - thanks so much @melissawm for doing all the hard work in merging stuff between repos! Things seem to be working as expected.

We had mentioned sphinx themes in today's meeting - the two that I have the most familiarity with are the [sphinx-book-theme](https://github.com/executablebooks/sphinx-book-theme) from the [executablebooks](https://github.com/executablebooks) project, and the [pydata-sphinx-theme](https://github.com/pandas-dev/pydata-sphinx-theme) on which the former is based. I only mention them because I know the book theme at least is explicitly designed with notebooks in mind. I just figured I'd mention them here since I brought up these themes in the meeting.

bjnath:
Trying to build the PR, `make html` fails with:
```
Extension error:
Could not import extension myst_nb (exception: No module named 'myst_nb')
```
I do have myst-parser-0.8.0 installed. Am I missing something else?

melissawm:
Did you do
```
pip install -r requirements.txt
```
before compiling? (This instruction is in the new README) This has all the dependencies listed and will install myst-nb automatically for you. If you are using conda you can install pip in your conda environment and then run the command above.

bjnath:
Thanks @melissawm  -- I see the new README now. Was able to build. 

- `tutorial-svd.ipynb` broke `make` because the demo ` s @ Vt`  fail is really a fail.  I propose commenting out the code cell and changing the text above it to read "Removing the `#` and executing" rather than "Executing."

- Thanks for the theme surgery! This style is more graceful, but I very much hope we don't use that headline typeface (Neuton).

rossbar:
> tutorial-svd.ipynb broke make because the demo s @ Vt fail is really a fail. I propose commenting out the code cell and changing the text above it to read "Removing the # and executing" rather than "Executing."

The cell in question here is *intended* to raise an exception to illustrate something about the shapes of the arrays returned by `np.linalg.svd`, so it should not be commented out. MyST-NB provides a way to prevent the execution kernel from terminating when an exception is expected - you can add a `raises-exception` tag to the cell in question. I should've mentioned that these were the only modifications I made to the two tutorials: both the cs231n and svd tutorials had cells that intentionally raised exceptions for demonstration purposes. It seems to me though that this change was successfully ported over in 8e60be1. Maybe you have an old version of the original svd tutorial (a merge conflict?) and the one in 8e60be1 isn't being picked up?

bjnath:
> intended to raise an exception

That was clear.

>prevent the execution kernel from terminating when an exception is expected

While that's a nice MyST-NB feature, we have use cases that don't involve MyST-NB. My wording keeps the intention clear and also works in real notebooks.

> an old version

In a clean directory that feature indeed works, and the problem would be solved if we were only serving web pages.

bjnath:
In case I wasn't clear, the "commented-out" cell shows up the same as before, only with a `#` prepended.

image


We know from math class that the shapes won't work, and we tell them that it won't work in NumPy either, and we tell them a `ValueError` results, so I think the point is pretty much driven home.

Notebook users who want to see the fireworks can remove the `#`.

rossbar:
> While that's a nice MyST-NB feature, we have use cases that don't involve MyST-NB. My wording keeps the intention clear and also works in real notebooks.

This is a general feature of most of the sphinx extensions (whether it's jupyter-sphinx, MyST-NB, nbsphinx, etc.) that deal with executing notebooks as part of the site build process. You're absolutely right that this shouldn't (and doesn't) affect the notebook content nor the interactive execution when users download the notebooks and run them in their own sessions. However for auto-generating the html from the notebook source files, it is often necessary to get around thorny issues like this (or add other features like skipping cell execution, running cells only in certain contexts [e.g. testing], etc.). The notebooks provide a nice mechanism to add these features either by adding metadata or tags to individual cells. In this case (and the instance in the cs231n tutorial that I modified as well), I think it's better for the rendered html to retain the actual exception, as the traceback is useful.

bjnath:
> the traceback is useful

If a traceback is useful, let's put it in a text cell.  

Think about whom we're serving: We're providing tutorial notebooks for beginning users.  They need notebooks they can execute simply. 

If you're a new notebook user who's pressed Run Cells, you'll be baffled that execution stops in the middle of the notebook and cells stop being filled in. 

Tutorials are the wrong use case for deluxe HTML features. Even if the tools can dance around land mines, we shouldn't be serving notebooks with land mines.  

As a point of policy,  let's offer notebooks that Just Work.

melissawm:
Thanks @rossbar for the suggestion, the theme seems to work great! This is ready for review (and hopefully we can launch this soon!)

conf.py
```diff
@@ -0,0 +1,79 @@
...
+# release = '0.0.1-dev'
+
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
+    "myst_nb",
```
eric-wieser, line 34: What was the tradeoff between `myst_nb` and `nbsphinx`? I've other projects which use the latter, but if the former was found to be the better choice then I might be interested in swapping over my projects too.

rossbar: I'm not an expert here but I can at least try to provide a little background. `myst-nb` is one part of the larger [executablebooks project](https://github.com/executablebooks) that aims to improve the tooling around the generation of scientific content in a way that supports many input formats. I have not used `nbsphinx` extensively, but the two main reasons I really like `myst-nb` are:
 1. Support for a [text-based notebook format](https://myst-nb.readthedocs.io/en/latest/use/markdown.html)
 2. Built-in support for [caching notebook execution](https://myst-nb.readthedocs.io/en/latest/use/execute.html)

The first point probably isn't so relevant if you already have a bunch of content in traditional `.ipynb` form (though [jupytext supports bi-directional conversion between `ipynb` and `mystnb` format](https://jupytext.readthedocs.io/en/latest/formats.html#myst-markdown), making it easy to try out the text-based format with your existing content).

Execution caching is another nice feature, especially if you have a lot of different source notebooks/files. Note that [jupyter-cache](https://github.com/executablebooks/jupyter-cache) is a separate project, so it's probably possible to get it working with `nbsphinx`. However, it's included by default with `myst-nb` (since it's also part of executablebooks).

There are other reasons as well (`nbsphinx` has an external dependency on `pandoc` whereas (AFAICT) `myst-nb` has no external [non-Python] dependencies). This is by no means an exhaustive list of the differences, just some of the reasons that I proposed `myst-nb` for this project.

melissawm: On the rendering side, I've also noticed that `myst-nb` is a bit more polished, whereas there were some formatting issues with `nbsphinx` (these are cosmetic details but I think those are important here).

melissawm:
Hi all, I'm merging this in the interest of moving things forward. This is not necessarily the best setup, and we can still rediscuss this, but right now I think it would be best to have something up and running and get all the details right later.

PR #15, opened 2020-07-21
Alternative I/O how-to

bjnath:
Taking a different approach than PR #14 for the I/O how-to. This is a leaner style.

The draft glosses over pandas; maybe it deserves more play. Among Stack Overflow questions tagged `numpy`, [Convert pandas dataframe to NumPy array](https://stackoverflow.com/questions/13187778/convert-pandas-dataframe-to-numpy-array) is among the top 10 votegetters.

Also, I wasn't sure whether additional or different formats should be listed for exchanging data with other tools.

### DIFF
--- a/howto_io.md
+++ b/howto_io.md
@@ -1,29 +1,36 @@
 # How-to: NumPy I/O
-- [Read a .csv or other text file with no missing values](#read-a-csv-or-other-text-file-with-no-missing-values)
-- [Read a .csv or other text file with missing values](#read-a-csv-or-other-text-file-with-missing-values)
-  * [Non-whitespace delimiters](#non-whitespace-delimiters)
-    + [Masked-array output](#masked-array-output)
-    + [Array output](#array-output)
-    + [Array output, specified fill-in value](#array-output-specified-fill-in-value)
-  * [Whitespace-delimited](#whitespace-delimited)
-- [Read a file in .npy or .npz format](#read-a-file-in-npy-or-npz-format)
-- [Write to a file to be read back by NumPy](#write-to-a-file-to-be-read-back-by-numpy)
-  * [Binary](#binary)
-  * [Human-readable](#human-readable)
-  * [Large arrays](#large-arrays)
-- [Read an arbitrarily formatted binary file ("binary blob")](#read-an-arbitrarily-formatted-binary-file-binary-blob)
-- [Write or read very large arrays](#write-or-read-very-large-arrays)
-- [Write files for reading by other (non-NumPy) tools](#write-files-for-reading-by-other-non-numpy-tools)
-- [Write or read a JSON file](#write-or-read-a-json-file)
-- [Save/restore using a pickle file](#saverestore-using-a-pickle-file)
-- [Convert from a pandas DataFrame to a NumPy array](#convert-from-a-pandas-dataframe-to-a-numpy-array)
-- [Save/restore using `numpy.tofile` and `numpy.fromfile`](#saverestore-using-numpytofile-and-numpyfromfile)
+- [1 Reading text files](#1-reading-text-files)
+  * [1.1 Read a CSV or other text file with no missing values](#11-read-a-csvhttpsenwikipediaorgwikicomma-separated_values-or-other-text-file-with-no-missing-values)
+  * [1.2 Read a CSV or other text file with missing values](#12-read-a-csvhttpsenwikipediaorgwikicomma-separated_values-or-other-text-file-with-missing-values)
+    + [1.2.1 Non-whitespace delimiters](#121-non-whitespace-delimiters)
+      - [1.2.1.1 Masked-array output](#1211-masked-array-output)
+      - [1.2.1.2 Array output](#1212-array-output)
+      - [1.2.1.3 Array output, specified fill-in value](#1213-array-output-specified-fill-in-value)
+    + [1.2.2 Whitespace-delimited](#122-whitespace-delimited)
+- [2 Read a file in .npy or .npz format](#2-read-a-file-in-npy-or-npz-format)
+- [3 Write to a file to be read back by NumPy](#3-write-to-a-file-to-be-read-back-by-numpy)
+  * [3.1 Binary](#31-binary)
+  * [3.2 Human-readable](#32-human-readable)
+  * [3.3 Large arrays](#33-large-arrays)
+- [4 Read an arbitrarily formatted binary file ("binary blob")](#4-read-an-arbitrarily-formatted-binary-file-binary-blob)
+- [5 Write or read large arrays](#5-write-or-read-large-arrays)
+- [6 Write files for reading by other (non-NumPy) tools](#6-write-files-for-reading-by-other-non-numpy-tools)
+- [7 Write or read a JSON file](#7-write-or-read-a-json-file)
+- [8 Save/restore using a pickle file](#8-saverestore-using-a-pickle-file)
+- [9 Convert from a pandas DataFrame to a NumPy array](#9-convert-from-a-pandas-dataframe-to-a-numpy-array)
+- [10 Save/restore using `numpy.tofile` and `numpy.fromfile`](#10-saverestore-using-numpytofile-and-numpyfromfile)
+- [11 See all NumPy's I/O routines](#11-see-all-numpys-io-routines)
 
-## Read a .csv or other text file with no missing values
+<a name="1-reading-text-files"></a>
+## 1 Reading text files
+
+<a name="11-read-a-csvhttpsenwikipediaorgwikicomma-separated_values-or-other-text-file-with-no-missing-values"></a>
+### 1.1 Read a [CSV](https://en.wikipedia.org/wiki/Comma-separated_values) or other text file with no missing values
 
 Use [numpy.loadtxt](https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html).
 
-## Read a .csv or other text file with missing values
+<a name="12-read-a-csvhttpsenwikipediaorgwikicomma-separated_values-or-other-text-file-with-missing-values"></a>
+### 1.2 Read a [CSV](https://en.wikipedia.org/wiki/Comma-separated_values) or other text file with missing values
 
 Use
 [numpy.genfromtxt](https://numpy.org/doc/stable/user/basics.io.genfromtxt.html).
@@ -34,7 +41,8 @@ masking out missing
 values (if ``usemask=True``) or will fill in the missing value with the value
 specified in ``filling_values`` (default is ``np.nan`` for float, -1 for int).
 
-### Non-whitespace delimiters
+<a name="121-non-whitespace-delimiters"></a>
+#### 1.2.1 Non-whitespace delimiters
 
 ```
 $ cat csv.txt
@@ -42,7 +50,8 @@ $ cat csv.txt
 4,, 6
 7, 8, 9
 ```
-#### Masked-array output
+<a name="1211-masked-array-output"></a>
+##### 1.2.1.1 Masked-array output
 ```
 >>> np.genfromtxt("csv.txt", delimiter=",", usemask=True)
 masked_array(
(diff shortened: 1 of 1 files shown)


howto_io.md
```diff
@@ -0,0 +1,268 @@
+# How-to: NumPy I/O
```
mattip, line 1: Needs a two sentence intro

howto_io.md
```diff
@@ -0,0 +1,268 @@
...
+  * [Large arrays](#large-arrays)
+- [Read an arbitrarily formatted binary file ("binary blob")](#read-an-arbitrarily-formatted-binary-file-binary-blob)
+- [Write or read very large arrays](#write-or-read-very-large-arrays)
+- [Write files for reading by other (non-NumPy) tools](#write-files-for-reading-by-other-non-numpy-tools)
+- [Write or read a JSON file](#write-or-read-a-json-file)
+- [Save/restore using a pickle file](#saverestore-using-a-pickle-file)
+- [Convert from a pandas DataFrame to a NumPy array](#convert-from-a-pandas-dataframe-to-a-numpy-array)
+- [Save/restore using `numpy.tofile` and `numpy.fromfile`](#saverestore-using-numpytofile-and-numpyfromfile)
+
+## Read a .csv or other text file with no missing values
```
mattip, line 22: Should define the acronym and explain in ten words what missing values are

howto_io.md
```diff
@@ -0,0 +1,268 @@
...
+Use
+[numpy.genfromtxt](https://numpy.org/doc/stable/user/basics.io.genfromtxt.html).
+
+``numpy.genfromtxt`` will either return a
+[masked array](https://numpy.org/doc/stable/reference/maskedarray.generic.html)
+masking out missing
+values (if ``usemask=True``) or will fill in the missing value with the value
+specified in ``filling_values`` (default is ``np.nan`` for float, -1 for int).
+
+### Non-whitespace delimiters
```
mattip, line 37: I think this should all be under a top-level "Reading formatted text files", with an introduction narrative including the CSV acronym,  loadtxt and genfromtext, with a sentence about when to use each. Then this section's title should be "Missing Values: CSV files" and the next section "Missing Values: other formats"

howto_io.md
```diff
@@ -0,0 +1,268 @@
...
+44^I ^I6
+7^I888^I9
+
+>>> np.genfromtxt("tabs.txt",delimiter="\t",missing_values=" +")
+array([[  1.,   2.,   3.],
+       [ 44.,  nan,   6.],
+       [  7., 888.,   9.]])
+```
+
+## Read a file in .npy or .npz format
```
mattip, line 146: Reverse the order of write and read, so you can say that "load can read files generated by any of save, savez, or savez_compressed"

howto_io.md
```diff
@@ -0,0 +1,268 @@
...
+    ("block_align", "<u2"),
+    ("bits_per_sample", "<u2"),
+    ("data_id", "S4"),
+    ("data_size", "u4"),
+    #
+    # the sound data itself cannot be represented here:
+    # it does not have a fixed size
+   ])
+
+data = np.load(f,dtype=wave_header_dtype)
```
mattip, line 213: The example (at the link http://scipy-lectures.org/advanced/advanced_numpy/index.html#example-reading-wav-files) actually uses np.fromfile, no?

howto_io.md
```diff
@@ -0,0 +1,268 @@
...
+
+Itamar Turner-Trauring in
+[pythonspeed.com](https://pythonspeed.com/articles/mmap-vs-zarr-hdf5/) gives
+more details and describes tradeoffs among memmap, Zarr, and HDF5.
+
+* For **HDF5**, use [h5py](https://www.h5py.org/) or [PyTables](https://www.pytables.org/).
+* For **Zarr**, see [here](https://zarr.readthedocs.io/en/stable/tutorial.html#reading-and-writing-data).
+* For **NetCDF**, use [scipy.io.netcdf_file](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.netcdf_file.html).
+
+## Write files for reading by other (non-NumPy) tools
```
mattip, line 242: Not sure what this section contributes.

howto_io.md
```diff
@@ -0,0 +1,268 @@
...
+
+Formats for exchanging data with other tools include HDF5, Zarr, and NetCDF.
+
+## Write or read a JSON file
+
+NumPy arrays are not directly [JSON serializable](https://github.com/numpy/numpy/issues/12481).
```
mattip, line 248: Might be helpful to mention "since the JSON format only defines ints and floats, while NumPy has a rich set of dtypes"

howto_io.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Save/restore using a pickle file
+
+Not recommended, due to lack of security and portability.
+
+ * security: not secure against erroneous or maliciously constructed data
+ * portability: may not be loadable on different Python installations
+
+Use `np.save` and `np.load`, setting ``allow_pickle=False``.
```
mattip, line 257: Might mention that this is required for NumPy arrays with dtypes that include python objects.

howto_io.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Convert from a pandas DataFrame to a NumPy array
+`DataFrame.to_numpy()`
```
mattip, line 260: Is there a link in the Pandas documentation that the stackoverflow answers point to?

howto_io.md
bjnath: Thanks for taking time to review this.

> Is there a reason you chose not to present this as a Jupyter Notebook?

* It was a prototype to see if there was support for the PR.
* Unlike a tutorial, where the user is invited to work with the examples, this are rote steps; many are just links.
* To make this executable in a user environment (for instance, via binder), files used in the examples would be introduced as dependencies, plus libraries like Zarr that the user would need to install. 
* If it's just documentation and not user-runnable, nothing is added by making it a notebook; Markdown is easier to update and review.

> The file itself should be moved to the content directory

Since it isn't a notebook, I wanted to keep it clear of whatever CI / build mechanism is going on in the repo till I knew what would happen if an .md were thrown in.

Also the repo isn't yet organized to separately handle tutorials and how-tos.

> Maybe the IO section of this repo should have two documents: a tutorial and a how-to. PR gh-14 is more of a tutorial, this is more of a how-to.

I agree there's value in covering topics with both tutorials and how-tos.  gh-14 has a tutorial feel but doesn't qualify as a tutorial in its current form. A tutorial needs a single narrative thread; it can't be a catalog.

> Needs a two sentence intro
If the title is something like "How to: NumPy I/O", what would be added by an intro?

howto_io.md
bjnath: * Should define the acronym

Some acronyms need no introduction. We don't spell out ASCII. `.csv` is widely known even to people who have no idea what NumPy is, and will certainly be known to the person who needs to read a file with a .csv extension. Moreover, csv delimiters aren't always commas, so spelling it out makes it less clear.  

A compromise might be to link to the [Wikipedia article](https://en.wikipedia.org/wiki/Comma-separated_values). In that case I'd follow the Google doc style guide (which I didn't here) and call the format CSV,

* explain in ten words what missing values are

What wording do you think would be helpful? Won't it be clear from the how-to headings that a "missing value" can be either a field missing data so the column count runs short, or a field that contains a special marker indicating the data is unusable?  And won't users reading data files be painfully knowledgeable on the subject of missing data?

howto_io.md
bjnath: NumPy converts simpler types in other contexts, so the reader who sees this explanation may be skeptical. It's got to be more than this.  We do link to the issue, and having read it it just looks like a nontrivial problem NumPy hasn't chosen to solve.

howto_io.md
bjnath: I'd have put it in. There didn't seem to be any clear or uniform or authoritative-sounding answer, and I didn't know enough pandas to step in.

howto_io.md
bjnath: NumPy converts simpler types in other contexts, so the reader who sees this explanation may be skeptical. It's got to be more than this.  We do link to the issue, and having read it it just looks like a nontrivial problem NumPy hasn't chosen to solve.

howto_io.md
bjnath: I'd have put it in. There didn't seem to be any clear or uniform or authoritative-sounding answer, and I didn't know enough pandas to step in.

howto_io.md
bjnath: Thanks -- the tutorial focused on the dtype and didn't actually show the read routine, and I supplied the wrong one. Fixed.

howto_io.md
bjnath: Thanks, will add.

howto_io.md
bjnath: Done.

bjnath:
[Preview of updated file.](https://github.com/numpy/numpy-tutorials/blob/65ceb532992c0119e9978236cba9aec721d80170/howto_io.md)

mattip:
I disagree with the premise that a how-to has absolutely no narrative text. Could you point to another project that uses this kind of format? The skeleton page you link to above is too sparse.

bjnath:
> the premise that a how-to has absolutely no narrative text

What's making you think I believe this? I don't.  The bottom line is always, "Does it help the user?", and I'd be a charlatan if I let anything else come first.

> too sparse

I'm not certain what you're after. Is the suggestion that a page titled **How-to: NumPy I/O** in a section of NumPy documentation labeled **How-tos** needs an opening sentence "This is a how-to on NumPy I/O"?  

The page is "too sparse" if people can't find the information they're looking for. It's "just right" if they find the information quickly and can use it.  Do you disagree?

howto_io.md
mattip: It would set expectations and provide keywords to help discoverability. "NumPy has many methods to serialize ndarrays. This How-To will provide an overview of when to choose which one, based on the task at hand, with links for more information and examples. Since this is a How-To (link to the how-to explaination), it is purposefully abbreviated to be as short as possible."

howto_io.md
```diff
@@ -0,0 +1,304 @@
...
+- [7 Write or read a JSON file](#7-write-or-read-a-json-file)
+- [8 Save/restore using a pickle file](#8-saverestore-using-a-pickle-file)
+- [9 Convert from a pandas DataFrame to a NumPy array](#9-convert-from-a-pandas-dataframe-to-a-numpy-array)
+- [10 Save/restore using `numpy.tofile` and `numpy.fromfile`](#10-saverestore-using-numpytofile-and-numpyfromfile)
+- [11 See all NumPy's I/O routines](#11-see-all-numpys-io-routines)
+
+<a name="1-reading-text-files"></a>
+## 1 Reading text files
+
+<a name="11-read-a-csvhttpsenwikipediaorgwikicomma-separated_values-or-other-text-file-with-no-missing-values"></a>
```
mattip, line 27: A shorter anchor would be better

howto_io.md
```diff
@@ -0,0 +1,304 @@
...
+```
+Credit: [Pauli Virtanen](http://scipy-lectures.org/advanced/advanced_numpy/index.html)
+
+<a name="large_arrays">
+
+<a name="5-write-or-read-large-arrays"></a>
+## 5 Write or read large arrays
+
+Arrays too large to fit in memory can be treated like ordinary in-memory arrays using
+[numpy.mmap](https://numpy.org/doc/stable/reference/generated/numpy.memmap.html).
```
mattip, line 247: In light of issue numpy/numpy#16979 maybe add something along the lines of "also useful to reduce the number of data copies made via `large_array[some_slice] = np.load('path/to/small_array', mmap_mode='r')`

howto_io.md
mattip: Maybe save the users a search by linking directly to the pandas documentation of to_numpy either with an intersphinx `:ref:` (are we using sphinx here?) or directly to https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_numpy.html

howto_io.md
bjnath: Valid point. If I may, *The Zen of Anchors*:

> An anchor that works is better than a missing anchor.
> An anchor generated by software is better than one generated by hand.
> The look of an anchor is the sound of one hand clapping.

So yes, the anchor length is ridiculous, but since it isn't clear what format this doc will ultimately take, I'd like to set this consideration aside for now.

howto_io.md
bjnath: Great suggestion. Added.

bjnath:
> linking directly to the pandas documentation 

Thank you; that's the info I was hoping someone would help with.

> Someone else will have to comment on the markdown/notebook format and in which directory to put this file.

My hope is that no one will comment on either of these things until we've talked about content. I chose an arbitrary format so we could start the discussion on how-tos.  Let's first pick a direction, and following that have a profitable discussion on what format and location works best.

howto_io.md
bjnath: Opening with "NumPy has many methods to serialize ndarrays" is like answering "There are many ways to tell time" to someone who asks the time.

Just like the guy asking for the time, the reader's expectations are already in place: they have a how-to question and expect to find the answer. Our job is to meet the expectations or fail.

If it weren't obvious how to use the contents -- if, say, we were presenting a triangular mileage map -- we might want to spare a few words to explain it. But everyone can pattern-match on headings. 

image

howto_io.md
bjnath: Thank you; that's the info I was hoping someone would help with.

howto_io.md
mattip: The difference is that we are not the only person for miles around on a rural road being queried nor are we an atlas sitting on a dusty shelf. We are a  web-page on an internet full of other versions of the same information. If we wish to be noticed, we need to rise to the top of the heap: we need to be indexed and searchable and appear at the top of searches. An appropriate minimum of words helps that process. A bare list of links does not: it scores very low and will not be noticed.

bjnath: I might agree more if I could find a single useful keyword in:

> NumPy has many methods to serialize ndarrays. This How-To will provide an overview of when to choose which one, based on the task at hand, with links for more information and examples. Since this is a How-To (link to the how-to explaination), it is purposefully abbreviated to be as short as possible.

The "bare links" point to headings which are full of keywords.  Search engines have no trouble finding our api pages, which have no empty sentences propping them up.

melissawm:
As discussed in the Documentation Team meeting today, I think this can be merged and if there are outstanding issues we can open separate PR's for that. However, I think now is the time to convert this to a notebook. @bjnath can you do that?

bjnath:
I have real problems with the formatting.  Although I put a TOC on top, I think most people will still be scanning through the text. The how-to is written like a decision tree, and the typography of headings doesn't help you understand where in the tree you are. That's what I tried to solve with numbered headings. 

The lost-in-a-large-file-where-am-I problem is fixed if we make this an rst -- the right sidebar in the new theme shows exactly where you are.  Numbered headings and a manually added TOC are unnecessary. That is another reason I think how-tos should be rst files, not notebooks, as argued in #16.  

I do agree with Matti that we can't have numbered headings that aren't automatically generated. And I think that if we have numbered headings in one how-to, we probably have to commit to using them on every how-to.

melissawm:
Maybe both documents should then be moved to the main repo. The formatting is more appropriate and the content, too.

bjnath:
The guide has gone Sphinx: PR 17353

PR #16, opened 2020-07-25
DOC: How-to guide for how-tos

bjnath:
Writing guidelines for how-to docs.

### DIFF
--- a/how_to_how_to.md
+++ b/how_to_how_to.md
@@ -1,0 +1,69 @@
+# How to write a NumPy how-to
+
+How-tos get straight to the point. They either answer a focused question, or narrow down
+broad questions into a set of focused questions that the user can then choose among.
+
+## A stranger has asked for directions
+
+*"I need to fill my tank."*
+
+## Brief but explicit answer
+
+  * Three miles, take a right at Hayseed Road, it's on your left.
+
+Add useful details ("Hayseed Road", even though it's the only turnoff at three miles), but not irrelevant ones. Don't say "A lot of people have trouble finding fuel in this town." If you're on Route 3, don't also explain how to get there from Route 7. If there's interesting background (tutorial, explanation, reference, alternative approach), link to it so the user can discover it, but that information doesn't belong in a how-to.
+
+## Delegate
+
+  * Three miles, take a right at Hayseed Road, follow the signs.
+
+If the information is already in the docs, link to it -- don't repeat it. If the API doc is sufficiently clear, link directly to it after you've given the prefatory information ("Three miles, take a right").
+
+## If the question is broad, narrow and redirect it
+
+ *"I want to see the sights"*
+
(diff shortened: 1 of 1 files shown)


how_to_how_to.md
```diff
@@ -0,0 +1,69 @@
...
+
+  * Three miles, take a right at Hayseed Road, it's on your left.
+
+Add useful details ("Hayseed Road", even though it's the only turnoff at three miles), but not irrelevant ones. Don't say "A lot of people have trouble finding fuel in this town." If you're on Route 3, don't also explain how to get there from Route 7. If there's interesting background (tutorial, explanation, reference, alternative approach), link to it so the user can discover it, but that information doesn't belong in a how-to.
+
+## Delegate
+
+  * Three miles, take a right at Hayseed Road, follow the signs.
+
+If the information is already in the docs, link to it -- don't repeat it. If the API doc is sufficiently clear, link directly to it after you've given the prefatory information ("Three miles, take a right").
```
melissawm, line 20: nit: I had no idea what prefatory meant (loved learning about it!), maybe change it to a simpler word for the benefit of non-native English speakers?

melissawm:
Maybe @rossbar or @mattip would like to take a look?

how_to_how_to.md
8bitmp3: I really like this How-to guide for how-tos @bjnath - thank you!

As @melissawm said, maybe change it to a simpler word. What do you think about using "introductory"? (Ref: https://developers.google.com/style/translation#be-inclusive)

Cheers!

melissawm:
As discussed in the Documentation Team meeting, I'd like to merge this. However, this guide (just like the I/O How-to #15 ) needs to be in notebook format, which would make it easier for people to browse using Binder, for example.

bjnath:
Something we haven't discussed: While notebooks are perfect for tutorials, they're wrong for how-tos.

Tutorials are meant to be executable, so notebooks are great for them. 

How-tos aren't meant to be executable. You can see that in #15 --  there's nothing to execute there. It's just names of commands and pointers to other docs.  

Thus there's no value in putting a how-to in a notebook. It would consist of a single text cell.  

Two conclusions: 

1. If how-tos aren't notebooks, they probably belong not in this repo but with the main Sphinx docs.

2. If people nonetheless see a need for how-tos to be notebooks, it suggests we haven't converged on what a how-to is, and it's probably too soon to merge.

melissawm:
This makes sense. My thinking was that writing them in notebooks might make the barrier of entry a little smaller for people. However, if you're writing a how-to, maybe that barrier is not so relevant. In that case, this how-to should be submitted to the main repo instead of here.

bjnath:
Per discussion, the content is now on the Sphinx pages: https://numpy.org/devdocs/user/how-to-how-to.html

Issue #10, opened 2020-06-10
Make notebooks easily executable


bjnath:
We agreed at the [8 June 20 doc team meeting](https://hackmd.io/oB_boakvRqKR-_2jRV-Qjg#Notes-from-8-June-20) that interactivity is a vital feature of notebooks.

To ensure this, each tutorial will include:

1. A button to run the notebook in **Binder**
1. Also a button to run the notebook in **Colab**. Although we'd prefer that users choose Binder, Colab runs more quickly. We'd give users a choice.
1.  A button to easily **download** the notebook for local execution.

Per @rossbar all three of these can be included.

Noting this here as a requirement for new notebooks.

melissawm:
With the new format for the repository, I think this is more or less solved. The only remaining question is whether we should have a colab link. Personally, I think Binder fits our project better.

rgommers:
> Personally, I think Binder fits our project better.

I agree, I'd prefer to support the community-developed solution here. We don't need GPUs, so Binder fits the need - the slightly faster startup for me is not a good reason to add Colab.

Related: the second tutorial in this repo has "Colab" in the name in the README - that looks weird.

PR #17, opened 2020-08-04
Removed sphinx and added binder 

### LINKED ISSUE #5: Check out online rendering of notebooks (i.e. Binder)
It's important to make the notebooks easily executable online. It would be preferable to do it via [mybinder.org](https://mybinder.org/) unless there are performance issues (for example, if the notebooks take too long to render).

melissawm:
After discussions with the documentation team, we decided to make this repo simpler and remove sphinx, using only Binder for rendering notebooks.

This PR updates the configuration of the repo, removes unnecessary configuration files and includes gh-actions configurations to trigger rendering the notebooks (with binder) in issue/PR comments - [see more details here](https://mybinder.readthedocs.io/en/latest/howto/gh-actions-badges.html).

Closes #5 and #10.


.github/workflows/chatops-binder.yaml
```diff
@@ -0,0 +1,34 @@
+#./github/workflows/chatops-binder.yaml
```
rgommers, line 1: did you mean to name this file `Chatops` and use that name in this file?

rgommers:
Gave the button a quick try. Binder starts in the repo root. Can you make it start in `content/` instead? That's the only directory that users should be accessing right?

.github/workflows/chatops-binder.yaml
melissawm: Fixed :)

melissawm:
Hi @rgommers , just a ping to see if there's anything else left to do before we merge this :)

rgommers (approved):
LGTM, thanks Melissa!

rgommers:
It's not happy yet, may be because it's still building, but need to check on this later:

image

PR #19, opened 2020-08-13
DOC: Update SVD tutorial plots & gitignore

pritesh-shrivastava:
I've added `%matplotlib inline` command to the Jupyter notebook for SVD tutorial. Without this command, the given image`img` doesn't show in a Jupyter notebook until executing the cell twice. 
Besides, I've also added the code output for the SVD notebook as it makes the tutorial easier to follow !!
I've also added `.ipynb_checkpoints` into the `.gitignore` file under the header `# Editor temporary/working/backup files #`.


melissawm:
Hello, @pritesh-shrivastava ! Thanks for taking a look.

I agree that we should add `%matplotlib inline` , that should make the notebook more complete. For the outputs, it is our understanding that they should not be kept in the notebooks in this repo, as the notebooks should be executed by users as necessary (either locally, by cloning the repo, or using binder - see https://github.com/numpy/numpy-tutorials/pull/17). 

Note that there's a rendered version of this tutorial at https://numpy.org/devdocs/user/tutorial-svd.html

Let me know if you have any questions, and welcome!

pritesh-shrivastava:
Thanks @melissawm for the feedback. I've removed the notebook output in the SVD tutorial. 
I've still kept matplotlib inline command in the svd notebook & ipynb_checkpoints in the gitignore file.
Let me know if you want me to make any other changes to this PR.

Thanks again !

melissawm:
This looks good to me, thanks @pritesh-shrivastava !

PR #29, opened 2020-10-07
Added masked arrays tutorial (converted from main documentation).

melissawm:
Tests workflow and github action for nbval.


melissawm:
Ok, I'll figure out what's going on with the nbval action because it's not triggered (I think it has to do with the branch I'm pushing into).

melissawm:
Now that we're adding a link to NumPy tutorials in `numpy.org`, I'm thinking it would be wiser to add a link to NumPy tutorials instead of having these build with the main NumPy docs - so this tutorial should be moved here. Let me know what you think.

rossbar (approved):
Agreed. Thanks @melissawm!

PR #33, opened 2020-10-11
Tutorial: NumPy deep learning on MNIST from scratch

8bitmp3:
Hi @melissawm @mattip @bjnath @rgommers 🙌 GSoD

> This tutorial demonstrates how to build and train a simple 3-layer [feed-forward neural network](https://en.wikipedia.org/wiki/Feedforward_neural_network) with [backpropagation](https://en.wikipedia.org/wiki/Backpropagation) learning from scratch.

All feedback is welcome!

Thank you everyone 🙏
cc @iamtrask

---

> ### Prerequisites
> 
> The reader should have some knowledge of Python, NumPy array manipulation, and linear algebra. In addition, you should be familiar with main concepts of [deep learning](https://en.wikipedia.org/wiki/Deep_learning). 
> 
> To refresh the memory, you can take the [Python](https://docs.python.org/dev/tutorial/index.html) and [Linear algebra on n-dimensional arrays](https://numpy.org/doc/stable/user/tutorial-svd.html) tutorials. 
> 
> You are advised to read the [Deep learning](http://www.cs.toronto.edu/~hinton/absps/NatureDeepReview.pdf) paper published in 2015 by Yann LeCun, Yoshua Bengio, and Geoffrey Hinton, who are regarded as some of the pioneers of the field.
> 
> In addition to NumPy, you will be utilizing the following Python standard modules for data loading and processing: 
> - [`urllib`](https://docs.python.org/3/library/urllib.html) for URL handling.
> - [`request`](https://docs.python.org/3/library/urllib.request.html) for URL opening.
> - [`gzip`](https://docs.python.org/3/library/gzip.html) for gzip file decompression.
> - [`pickle`](https://docs.python.org/3/library/pickle.html) to work with the pickle file format.
> - [`sys`](https://docs.python.org/3/library/sys.html) for output printing. 
> 
>     as well as:
> - [matplotlib](https://matplotlib.org/) for data visualization.
> 
> This tutorial can be run locally in an isolated environment, such as [Virtualenv](https://virtualenv.pypa.io/en/stable/). You can use [Jupyter Notebook or JupyterLab](https://jupyter.org/install) to run each notebook cell. Don't forget to [set up NumPy](https://numpy.org/doc/stable/user/absolute_beginners.html#installing-numpy) and [matplotlib](https://matplotlib.org/users/installing.html#installing-an-official-release).
> 
> ### Table of contents
> 
> 1. Load the MNIST dataset
> 
> 2. Preprocess the dataset
> 
> 3. Build and train a small neural network from scratch
> 
> 4. Next steps
> 

---


bjnath:
This surely was many hours of effort -- thank you.

A few clarifications in the opening lines might help readers:

<br>

> This tutorial demonstrates how to build and train a simple 3-layer feed-forward neural network with backpropagation  

1.  Why pick this type of NN? 
2. +1 for linking to Wikipedia on feedforward and backpropagation.

<br>

> how to parse and load the dataset without the aid of external deep learning libraries

Isn't it the entire NN that's written without external libraries, not just parsing and loading?

<br>

> without the aid of external deep learning libraries

Why would I, the reader, want to do that?

<br>

> This type of machine or deep learning is often referred to as supervised learning.

This sentence parses ambiguously and the reference of "this" is unclear. Maybe not worth fixing -- leave it out.

<br>

> This tutorial was adapted from the work by Andrew Trask

Trask might appreciate a link to the original work, especially if it's his new book, rather than simply to his home page.

<br>

 > inspired by the teachings of Joel Grus

 1. It's not clear in the sentence who was inspired -- you or Trask. 
 2. If this were "inspired by the teachings of Gandhi," you could end it there, but not everyone knows Grus. Can you explain briefly?
3. "approach" might be a better word here than "teachings."

<br>

 > The deep learning model that you will construct with NumPy will learn from the MNIST dataset, which contains 60,000 training and 10,000 test images and corresponding labels of handwritten digits from 0 to 9. 

1. The emphasis is misplaced. The goal is to build an image classifier -- a program that will teach itself, from a set of examples, to recognize handwritten digits 0-9. 

2. The MNIST description won't be clear to an unacquainted reader. It's a collection of handwritten digits 0-9, with separate cheat sheets ("labels") that identify the digits correctly. Then you'd explain that there are separate sets for training the recognizer and for testing it once it's trained. And perhaps only then you'd mention there are 60,000 and 10,000 files.  <br>
  Alternatively, assume people who don't recognize MNIST will read the link, and just call MNIST "the standard [MNIST](https://en.wikipedia.org/wiki/MNIST_database) dataset."

3. The [Wikipedia MNIST reference](https://en.wikipedia.org/wiki/MNIST_database) might be more helpful here than LeCun's page. 

<br>

melissawm:
Hi @8bitmp3 , thank you for this tutorial, I really think this will be a nice addition and it is quite an effort to implement this without any pre-packaged library. From the text it is clear that you are in your element, too!

I have a few comments, mainly in the interest of making the document a bit simpler. I think some of the functions can be simplified or rewritten, and this way we can use less code-comments and more text. I believe this can make the whole document more fluid and easier to follow. Feel free to adapt or try different things if you think it doesn't fit your own narrative.

I'm focusing first on the start and some of the functions there, but I'll continue trying to bring ideas for the rest. In particular, I think "Neural network building blocks with NumPy" and "Model architecture summary" run a bit long - I think your explanations are spot on, but I'm worried that we'll lose track of the code at this point. 

From an educational point of view, I usually create variables or import libraries close to where they will actually be used, instead of in the beginning of the file. So you'll see that I did that for the first sections. 

Then, since we are using a notebook, we might as well make use of [literate programming](https://en.wikipedia.org/wiki/Literate_programming) ideas; namely, instead of using code comments, using actual text as much as possible. This made me rethink some of your functions and realize that, for the purpose of this document, they are really not necessary since those operations will only be performed once. So I tried rewriting the initial download/save/retrieve data without using functions at all. Let me know what you think and if you believe strongly those should be functions we can work on that. Also, some of the text might need re-working now, I just took some of the code comments and adapted but you may have better ideas for that :) Keep in mind that my suggestions are not meant to be the best solution - just another way of doing things. 

Here's my take on the document (not that I only touched the first section, and removed the need for `sys` later in the file).

https://github.com/melissawm/numpy-tutorials/blob/melissa-deep-learning/content/tutorial-deep-learning-on-mnist.ipynb

Overall, I'm pretty happy that we'll see this kind of content on the repo - so thanks again and let me know if there's anything I can help you with!

8bitmp3:
Thanks @bjnath for the awesome feedback.

> 1. Why pick this type of NN?
> ...
> The goal is to build an image classifier -- a program that will teach itself, from a set of examples, to recognize handwritten digits 0-9.
> ...
> 3. The [Wikipedia MNIST reference](https://en.wikipedia.org/wiki/MNIST_database) might be more helpful here than LeCun's page.

Changed the words around and refactored for clarity: 

```
This tutorial demonstrates how to build and train a simple 3-layer [feedforward neural network](https://en.wikipedia.org/wiki/Feedforward_neural_network) with [backpropagation](https://en.wikipedia.org/wiki/Backpropagation) from scratch with NumPy to recognize hardwritten digit images.

Your deep learning model — one of the most basic artificial neural networks that resembles the original [multi-layer perceptron](https://en.wikipedia.org/wiki/Multilayer_perceptron) — will learn to classify digits from 0 to 9 from the [MNIST](https://en.wikipedia.org/wiki/MNIST_database) dataset. 
```
 
> Isn't it the entire NN that's written without external libraries, not just parsing and loading?
> ...
> > without the aid of external deep learning libraries
> 
> Why would I, the reader, want to do that?

Agreed. Although it's not uncommon to use a deep learning library to add a popular dataset in a couple of lines of code, I've removed this anyway to avoid repetition.

> Trask might appreciate a link to the original work, especially if it's his new book, rather than simply to his home page.

Definitely. I also added a link to the repo and a reference to Trask's book - it's a must read.

I'm basing my third project on Grus' work, with his permission (a small [automatic differentiation](https://en.wikipedia.org/wiki/Automatic_differentiation) library with NumPy), so will mention him there instead of this project.

8bitmp3:
> I have a few comments, mainly in the interest of making the document a bit simpler. I think some of the functions can be simplified or rewritten

Definitely, thanks so much for the help @melissawm 🙌! I'll also soon simplify `one_hot_encoding()` as per [this solution](https://stackoverflow.com/questions/29831489/convert-array-of-indices-to-1-hot-encoded-numpy-array/29831596#29831596) based on feedback I got.

> I'm focusing first on the start and some of the functions there, but I'll continue trying to bring ideas for the rest. In particular, I think "Neural network building blocks with NumPy" and "Model architecture summary" run a bit long - I think your explanations are spot on, but I'm worried that we'll lose track of the code at this point.

Good point! I want to merge some of the stuff with the "compose section" (with the code). Simpler is better. Thanks so much! 

> From an educational point of view, I usually create variables or import libraries close to where they will actually be used, instead of in the beginning of the file. So you'll see that I did that for the first sections.

Yup, got it.

> Then, since we are using a notebook, we might as well make use of [literate programming](https://en.wikipedia.org/wiki/Literate_programming) ideas; namely, instead of using code comments, using actual text as much as possible. 

👍 I agree!

> So I tried rewriting the initial download/save/retrieve data without using functions at all. Let me know what you think and if you believe strongly those should be functions we can work on that. 

I think this is awesome. I mostly try to use the second person, but other than that 🙏.

Thank you 💯 !

melissawm:
Hi @8bitmp3 ! This is looking *really* good - I have a few comments but overall I think you did an awesome job. This is a tough subject and I do have some background in machine learning though - but I suspect this will also be the case for readers of this tutorial, so I think your links and references are right on target. 

I don't know if there's more simplification possible in the code - the main training loop is just as it should be. We could take some of the code-comments out but I'd like to see some other opinions on that. The only thing you might consider is building a function for the test/training loss+accurate prediction calculation. But I'm not sure that would be helpful for readers... For now my main comments will be on some typos and a few suggestions but nothing major. Thanks again for this!

content/tutorial-deep-learning-on-mnist.ipynb
```diff
@@ -0,0 +1,1017 @@
+{
```
melissawm, line 1: <p>We could use a link on "reverse-mode differentiation" if you have one.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

melissawm, line 1: <p>I would add to that a suggestion about conda, since it's become the standard for most scientific computing applications in Python.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

melissawm, line 1: <p>Here, and below, you mention reshaping the data to be <code>28x28</code>.  Can you add a note explaining why that is necessary?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

melissawm, line 1: <p>Suggestion: add a note about why it's necessary to use the gray colormap</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

melissawm, line 1: <p>Why do you use <code>interpolation='none'</code>  in this case, but not the previous one?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

melissawm, line 1: <p>nit: in the sentence " <code>training_labels</code>: labels corresponding to the training images of size 60,000." the reader might get the impression that the images are of size 60000. I'd change it to something like </p><p> </p><p><code>training_labels</code>:  array of size 60,000 corresponding to the labels of the training images</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

melissawm, line 1: <p>Can you add a citation/reference for the fact that smaller numbers in the matrix result in a faster operation?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

melissawm, line 1: <p>I think there is a word missing here: "In this section you will familiarize <strong>yourself</strong> with..."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='51'/>

melissawm, line 1: <ul><li>I think you meant matrix multiplication <strong>function</strong> <code>np.dot</code>  (instead of *module*)</li><li>In the activation function description I think you meant "inputs and outputs"  instead of "inputs and inputs" :)</li><li>"In this example, you will use a method called dropout (<a href="https://en.wikipedia.org/wiki/Dilution_(neural_networks))" target="_blank">dilution</a> that " (I think there is a formatting error here with the parenthesis?)</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='53'/>

melissawm, line 1: <p>The numbering for the items in this list seems to be off </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='62'/>

melissawm, line 1: <p>Can you add a sentence about why you are initializing the weights like this? Is that standard practice in machine learning, or is it really the choice of the user?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='65'/>

melissawm, line 1: <ul><li>Why do you test every 10 epochs? </li><li>Is it necessary to use <code>[i, i+1]</code>  for each item, or could <code>[i]</code>  do the same job?</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='67'/>

melissawm, line 1: <p>It would be really nice to close with a plot showing that your NN can identify an image that was not in the training set (I know the prints also show this, but just for a visual element).</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='68'/>

8bitmp3: Great idea @melissawm Also - I should pick a better loss function 🙃 The error rate it too high for this simple example, but if there's a will, there's a way!

A matplotlib visualization of the error rate falling and accuracy improving (hopefully) over time would also be super — and it's a common practice. Working on it!

8bitmp3: Absolutely @melissawm . I also added a note on "Xavier" initialization in the other tutorial - https://www.deeplearning.ai/ai-notes/initialization/ (it's also a very useful link for the reader).

8bitmp3: Will try to find a quality source. Also, for intuition, I can include some basic math (e.g. a 28x28 matrix times the number of examples times the number of neurons per layer, etc etc 👻 = a high number of parameters to learn.

8bitmp3: Good idea @melissawm. 

This is the closest reference I could find on Wikipedia:

```diff
... and backpropagation (
- reverse-mode
+ [reverse-mode](https://en.wikipedia.org/wiki/Automatic_differentiation#Reverse_accumulation)
differentiation)
...
```

8bitmp3: Done.

```diff
This tutorial can be run locally in an isolated environment, such as
[Virtualenv](https://virtualenv.pypa.io/en/stable/)
+ and [conda](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html)
...
```

8bitmp3: That's a great point - done. Here is the diff:

```diff
3. Extract and save the data to disk using NumPy for the first time to read the contents of each file as [`ndarrays`](https://numpy.org/doc/stable/reference/arrays.ndarray.html) and saving them into a dictionary.
+ Note that you need to unroll (reshape) the original 28x28-pixel images into 1D vectors
+ (of size 28 times 28 or 784) — it is the expected shape of the input for your neural
+ network's first layer and the number of input neurons per image sample.
```

(without the 80-char limit)

8bitmp3: Thanks for this suggestion. Here are the changes:

```diff
# reshape to (28, 28) to have a valid shape for displaying purposes.
...
- # Take the 60,000th image (index 59999) from the training set, reshape to (28x28) if necessary.
+ # Take the 60,000th image (index 59,999) from the training set,
+ # reshape from (784, ) to (28, 28) to have a valid shape for displaying purposes.
...
- # Set the color mapping to grayscale.
+ # Set the color mapping to grayscale to have a black background.
```

8bitmp3: Good point, makes no difference (due to the size of the images?) - removed it :)

8bitmp3: Nice catch.

```diff
 - `training_images`: flattened training images of size 60,000x784.
- - `training_labels`: labels corresponding to the training images of size 60,000.
+ - `training_labels`: array of size 60,000 corresponding to the labels of the training set.
 - `test_images`: flattened test images of size 60,000x784.
- - `test_labels`: labels corresponding to the training images of size 60,000.
+ - `test_labels`: array of size 10,000 corresponding to the labels of the test set.
```

8bitmp3: I think the Wiki page is "looking for citation" :) But I know of a similar technique that expands on the idea further (batch norm). The research paper mentions standardization as a way to speed up training, because the incoming data that the network has to learn from has the same distribution as the one before it:

```diff
...
- - _Normalizing_ the image data.
+ - _Normalizing_ the image data: a [feature scaling]
+ (https://en.wikipedia.org/wiki/Feature_scaling#Application) procedure that can
+ speed up the neural network training process by standardizing the
+ [distribution of your input data](https://arxiv.org/pdf/1502.03167.pdf).
...
- > **Note:** Since the numbers inside matrices will be smaller, it should also make it faster for the neural network to learn from that data.
```

8bitmp3: Got it!

8bitmp3: - Good catch :) Also changed "`np.random.random()` function" and "`np.random.randint()` function".
- Nice one. Changed to "inputs and outputs"
- There's a weird rendering issue in some notebook viewers - maybe because the URL ends with a bracket `[...](https://en.wikipedia.org/wiki/Dilution_(neural_networks))`. Changed it to:

```diff
- ... dropout ([dilution](https://en.wikipedia.org/wiki/Dilution_(neural_networks))) ...
+ dropout — [dilution](https://en.wikipedia.org/wiki/Dilution_(neural_networks)) — ...
```

8bitmp3: The rendering here is strange too, because I'm using `3. ... - _Learning rate_... - _Epochs (...)_` I added tabs, just in case:

```diff
3. Set certain default values of [hyperparameters](https://en.wikipedia.org/wiki/Hyperparameter_(machine_learning)), such as:
- - [_Learning rate_](https://en.wikipedia.org/wiki/Learning_rate): ...
+     - [_Learning rate_](https://en.wikipedia.org/wiki/Learning_rate): ...
...
```

8bitmp3: A very good question. I added this to the explainer section. It's at the heart of the algorithm, with it being a stochastic method.

```diff
 - _Weights_: These are important adjustable parameters that the neural network 
 fine-tunes by forward and backward propagating the data. 
+ They are optimized through a process called 
+ [gradient descent](https://en.wikipedia.org/wiki/Stochastic_gradient_descent). 
+ Before the model training starts, the weights are randomly initialized with 
+ NumPy's `np.random.random()` function.
    
    The optimal weights should produce the highest prediction accuracy and the lowest error on the training and test sets. 
- Before the model training starts, the weights are randomly initialized with 
- NumPy's `np.random.random()` function.
```

8bitmp3: I'm not 100% sure how to do "inference" with a trained model to classify images, since I don't think we have the [Saved Model](https://www.tensorflow.org/guide/saved_model) or [checkpoint saving](https://www.tensorflow.org/guide/checkpoint) capability in the simple example. But we can do the following:

TODO: 

- [ ] 1. We can add graphs with `matplotlib` for: 
  - The error rate (loss) for training and test sets (y) per epochs (x) (it should be gradually going down).
  - The accuracy rate for training and test sets (y) per epochs (x) (it should be gradually going up).
    - Need to add a few lines to save the loss and accuracy (prediction) data at each epoch by appending them separate lists (for storage).
- [ ] 2. Add a simple image to help new users:
  - Showing a loop: inputs as vectors -> forward pass through layers -> output (prediction) + loss calculation -> backpropagation with gradient descent -> weight update.

Other improvement suggestions:
- [ ] Use MSE or some simple loss function formula instead of the current one.
- [ ] Try Fashion MNIST - a similar dataset but more fun.

8bitmp3: - I think in the original gist, you test (evaluate) the model every 10 epochs because it should have more "knowledge" to be tested after the 10th training iteration. I've tried testing at every epoch and noticed the results were OK in this simple example. It's reflected in the new code.
- `[i:i+1]` is necessary here (doesn't seem to work without it)

8bitmp3: @melissawm PTAL 😃 Thank you!

melissawm: <p>nit: I'd say virtualenv <em>or</em> conda, but that's just the mathematician in me talking :)</p>

melissawm: <p>I think the references are good, but then I'd remove this note about smaller numbers causing faster learning, unless we have a good citation. I'm not a specialist in this so I don't necessarily see the connection :)</p>

melissawm: <p>The regularization link seems to be missing a parenthesis at the end. Thanks for the other fixes :)</p>

melissawm: <ul><li>Why do you print "Epoch" + str(j+10)? I think there is no need to add 10 to this result - am I missing something?</li><li>Ok I got the reason for the indexing but this code could be simplified as follows:</li></ul><pre class="ql-syntax" spellcheck="false"># To store training and test set losses and accurate predictions
# for visualization.
store_training_loss = []
store_training_accurate_pred = []
store_test_loss = []
store_test_accurate_pred = []

# This is a training loop.
# Run the learning experiment for a defined number of epochs (iterations).
for j in range(epochs):
&nbsp;&nbsp;&nbsp; # Set the initial loss/error and the number of accurate predictions to zero.
&nbsp;&nbsp;&nbsp; training_loss = 0.0
&nbsp;&nbsp;&nbsp; training_accurate_predictions = 0
&nbsp;&nbsp; &nbsp;
&nbsp;&nbsp;&nbsp; # For all images in the training set, perform a forward pass
&nbsp;&nbsp;&nbsp; # and backpropagation and adjust the weights accordingly.
&nbsp;&nbsp;&nbsp; for i in range(len(training_images)):
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # Forward propagation/forward pass:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 1. The input layer:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; Initialize the training image data as inputs.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_0 = training_images[i]
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 2. The hidden layer:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; Take in the training image data into the middle layer by
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; matrix-multiplying it by randomly initialized weights.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_1 = np.dot(layer_0, weights_1)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 3. Pass the hidden layer's output through the ReLU activation function.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_1 = relu(layer_1)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 4. Define the dropout function for regularization.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; dropout_mask = np.random.randint(0, high=2, size=layer_1.shape)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 5. Apply dropout to the hidden layer's output.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_1 *= dropout_mask * 2
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 6. The output layer:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; Ingest the output of the middle layer into the the final layer
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; by matrix-multiplying it by randomly initialized weights.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; Produce a 10-dimension vector with 10 scores.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_2 = np.dot(layer_1, weights_2)

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # Backpropagation/backward pass:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 1. Measure the training error (loss function) between the actual
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; image labels (the truth) and the prediction by the model.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; training_loss += np.sum((training_labels[i] - layer_2) ** 2)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 2. Increment the accurate prediction count.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; training_accurate_predictions += int(np.argmax(layer_2) == np.argmax(training_labels[i]))
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 3. Differentiate the loss function/error.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_2_delta = (training_labels[i] - layer_2)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 4. Propagate the gradients of the loss function back through the hidden layer.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_1_delta = np.dot(weights_2, layer_2_delta) * relu2deriv(layer_1)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 5. Apply the dropout to the gradients.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_1_delta *= dropout_mask
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 6. Update the weights for the middle and input layers
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; by multiplying them by the learning rate and the gradients.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; weights_2 += learning_rate * np.outer(layer_1, layer_2_delta)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; weights_1 += learning_rate * np.outer(layer_0, layer_1_delta)
&nbsp;&nbsp; &nbsp;
&nbsp;&nbsp;&nbsp; # Store training set losses and accurate predictions.
&nbsp;&nbsp;&nbsp; store_training_loss.append(training_loss)
&nbsp;&nbsp;&nbsp; store_training_accurate_pred.append(training_accurate_predictions)

&nbsp;&nbsp;&nbsp; # Evaluate on the test set:
&nbsp;&nbsp;&nbsp; # 1. Set the initial error and the number of accurate predictions to zero.
&nbsp;&nbsp;&nbsp; test_loss = 0.0
&nbsp;&nbsp;&nbsp; test_accurate_predictions = 0
&nbsp;&nbsp; &nbsp;
&nbsp;&nbsp;&nbsp; # 2. Start testing the model by evaluating on the test image dataset.
&nbsp;&nbsp;&nbsp; for i in range(len(test_images)):
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 1. Pass the test images through the input layer.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_0 = test_images[i]
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 2. Compute the weighted sum of the test image inputs in and
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; pass the hidden layer's output through ReLU.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_1 = relu(np.dot(layer_0, weights_1))
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 3. Compute the weighted sum of the hidden layer's inputs.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; #&nbsp;&nbsp;&nbsp; Produce a 10-dimensional vector with 10 scores.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; layer_2 = np.dot(layer_1, weights_2)

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 4. Measure the error between the actual label (truth) and prediction values.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; test_loss += np.sum((test_labels[i] - layer_2) ** 2)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; # 5. Increment the accurate prediction count.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; test_accurate_predictions += int(np.argmax(layer_2) == np.argmax(test_labels[i]))

&nbsp;&nbsp;&nbsp; # Store test set losses and accurate predictions.
&nbsp;&nbsp;&nbsp; store_test_loss.append(test_loss)
&nbsp;&nbsp;&nbsp; store_test_accurate_pred.append(test_accurate_predictions)

&nbsp;&nbsp;&nbsp; # 3. Display the error and accuracy metrics in the output.
&nbsp;&nbsp;&nbsp; print("\n" + \
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; "Epoch: " + str(j) + \
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; " Training set error:" + str(training_loss/ float(len(training_images)))[0:5] +\
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; " Training set accuracy:" + str(training_accurate_predictions/ float(len(training_images))) +\
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; " Test set error:" + str(test_loss/ float(len(test_images)))[0:5] +\
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; " Test set accuracy:" + str(test_accurate_predictions/ float(len(test_images))))
</pre><p>Let me know if this makes sense to you. There was nothing wrong with the original code - I just think it is easier to avoid transposing if you can, and using 1d arrays instead of 2d. </p>

content/tutorial-deep-learning-on-mnist.ipynb
```diff
@@ -0,0 +1,1067 @@
...
```
melissawm, line 1: <p>I think it might be better to leave all credits at the bottom of the document.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

melissawm, line 1: <p>Item 4 is missing now :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='64'/>

melissawm, line 1: <p>The plots are great, I think that closes the tutorial nicely!</p><p> </p><p>Tip: you can add <code>plt.tight_layout()</code>  to improve on the spacing between the subplots - at least locally I had the title and x-axis label of the subplots superimposed without using <code>tight_layout</code>.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/numpy/numpy-tutorials/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='70'/>

bjnath:
> In this section, you will download the zipped MNIST dataset files originally stored in Yann LeCun's website. Then, you will transform them into 4 files of NumPy array type using built-in Python modules. Finally, you will split the arrays into training and test sets.

It would be useful to be explicit about the endpoint. That way readers can just read the first lines and skip down to the next heading.

> The first step is mostly bookkeeping. You will create four ndarrays as inputs to the neural network:
> ```
> 
> x_train, y_train, x_test, y_test = (mnist_dataset["training_images"],
>                                     mnist_dataset["training_labels"],
>                                     mnist_dataset["test_images"],
>                                     mnist_dataset["test_labels"])> ```
> ```
> The MNIST data as stored on [Yann LeCun's website](http://yann.lecun.com/exdb/mnist/)  consists of four compressed files that you'll read into the ndarrays.
> ...[Pick up the bulleted list]

bjnath:
> Extract and save the data to disk using NumPy for the first time to read the contents of each file as ndarrays and saving them into a dictionary.

We're not really saving to disk here -- that implies we're writing files. How about
> NumPy can decompress the 4 files on the fly and create 4 ndarrays, which we save in a dictionary.

bjnath:
```
> for name in filename[:2]:
>...
> for name in filename[-2:]:
```
Comments would make it a little clearer:
```
> # Images
> for name in filename[:2]:
>...
> # Labels
> for name in filename[-2:]:
```

bjnath:
> Split the data into training and test sets using the standard notation of and call the training and test set of images as x_train and x_test, and the set of labels as y_train and y_test:

Not clear what "standard notation" is referring to. How about:

> Split the data into training and test sets using the standard notation of `x` for data and `y` for labels, calling the training and test images x_train and x_test, and the labels y_train and y_test:

bjnath:
> To make sure you have the right data, print the shape of the sets:

If you were really making sure, you'd code a test, not a print statement. And you should say what the user should expect to see.

> You can confirm that the shape of the image arrays is xxx and label arrays is yyy:

bjnath:
> (Optional) You can inspect some samples of the image data with matplotlib and image labels:

'Optional' not needed. This step follows another optional step, looking at the shapes, so it can read:

> And you can inspect some images using matplotlib:

bjnath:
> You have loaded the MNIST dataset and saved it to disk as 4 files:

I apologize if I've misread the Python, but have we actually created 4 files? It looks like we have a dictionary of 4 arrays.

bjnath:
> Note: Data preparation in deep learning is a very challenging process...

In general, tutorials should not include side comments in the main text.

bjnath:
```
sample = 1000
training_images = x_train[0:sample] / 255
test_images = x_test.reshape(len(x_test), 28*28) / 255
```
1. Not clear (to me, anyway) why the test and training images are being treated differently.
2. I would call loud attention to the fact -- surprising to newcomers and characteristic of NumPy -- that the division is *not* truncating the result into uint8's (which would leave zeros everywhere) but silently *promoting* the dtype into uint64, and that casting explicitly is not necessary.
3. Don't put the note underneath, and don't use "Note" -- include it in the "Normalize the arrays..." text.

bjnath:
> Since your dataset is already represented by arrays containing 8-bit integers, let's convert their dtype to floating points, such as float64 — 64-bit (or double-precision) — by doing the following:
> 
> * Normalizing the image data: a feature scaling procedure that can speed up the neural network training process by standardizing the distribution of your input data.
> 
> * One-hot/categorical encoding of the image labels.
> 
This is wordy and doesn't really explain why we're doing this. I'd just go with something like

> We make two modifications that are conventional for neural networks:
> * [Rescale](https://en.wikipedia.org/wiki/Feature_scaling#Application) the data so it takes values from 0 to 1
> * Encode the labels as [one-hot/categorical](https://en.wikipedia.org/wiki/One-hot) values

bjnath:
> Note: You can also check that normalization was successful...

Saying "note" isn't necessary; the note sections are typographically incongruous; and Google style calls this kind of usage a placeholder and [discourages](https://developers.google.com/style/tone) it.

bjnath:
> Convert the image data to floating points

I can see the justification for saying "floating points" ("Convert the sheds to bungalows"), but I think it's more idiomatic to say "floating point" without the s ("Convert the tickets to business class."). Others can weigh in.

bjnath:
> Use the newly created function to one-hot encode the image labels and assign the train and test label data variables — y_train and y_test — to training_labels and test_labels, respectively:

Suggested:

> Encode the labels and assign the values to new variables:

bjnath:
> (Optional) Visualize a few samples of the one-hot encoded labels:
> ...and compare them to the original labels:

Suggested:
> Examine a few encoded labels
>...and compare to the originals:

bjnath:
> This type of machine or deep learning is often referred to as supervised learning

It's unclear which part of the preceding 120-word description "this type" refers to. Supervised learning means learning from labeled examples. If you want to explain what supervised learning is, say that. My suggestion, though, is to drop the sentence.

bjnath:
> will learn to classify digits from 0 to 9

Suggested:

> will learn to identify handwritten digits from 0 to 9

bjnath:
> This tutorial demonstrates how to build a simple feedforward neural network (with one hidden layer) and train it with backpropagation from scratch with NumPy to recognize hardwritten digit images.

I'd leave backpropagation out of the first line; you say later that it trains using forward propagation and backpropagation.
Also typo in "ha**r**dwritten."

> This tutorial demonstrates how to build a simple feedforward neural network (with one hidden layer) and train it with  from scratch with NumPy to recognize handwritten digits.

bjnath:
> The final output of the network is a vector of 10 scores — one for each handwritten digit image. 

It's unnecessary and confusing to introduce this detail here.  I suggest leaving it out.

bjnath:
> This tutorial was adapted from the work by Andrew Trask (with the author's permission).

Suggested:

> This tutorial was adapted with permission from Andrew Trask's [Grokking Deep Learning](https://www.manning.com/books/grokking-deep-learning), which teaches deep learning with NumPy.

And drop the reference in Prerequisites.

8bitmp3:
Thank you @bjnath for a thorough review again, I really appreciate it 👍 💯 !

---

> If you're explicit about what this section produces, readers can read just the first lines and skip down to the next heading. You could reword this as:

I prefer to have small introductions in the beginning of each section, in case some users want to know what the goal is. They can always skip it. Besides, It's like a short summary and it may also help people who are busy and/or scanning the stuff.

---

> This needs to be a little clearer about how the data in the files is formatted so we understand what this step is doing.  

> We're not really saving to disk here -- that implies we're writing files. 

Got it! This can help clear things up for new readers.

> "3. Decompress the 4 files and create 4 [`ndarrays`](https://numpy.org/doc/stable/reference/arrays.ndarray.html), saving them into a dictionary. Each original image is of size 28x28 and neural networks normally expect a 1D vector input; therefore, you also need to reshape the images by multiplying 28 by 28 (784)."
>
> {... code showing `.reshape(-1, 28*28)`}

---

> Comments would make it a little clearer:
> ```
> # Images
> for name in filename[:2]:
> ...
> # Labels
> for name in filename[-2:]:
> ```

Yup! Agree.

@melissawm Can you please double check the code and comments, just in case 👍 

---

> Split the data into training and test sets using the standard notation of x for data and y for labels, calling the training and test images x_train and x_test, and the labels y_train and y_test:

Good idea!

---

> If you were really making sure, you'd code a test, not a print statement. And you should say what the user should expect to see. How about
> You can confirm that the shape of the image arrays is xxx and label arrays is yyy:

Super. Very good point. I changed it to the following:

> "5. You can confirm that the shape of the image arrays is `(60000, 784)` and `(10000, 784)` for training and test sets, respectively, and the labels — `(60000,)` and `(10000,)`:"

---

> And you can inspect some images using matplotlib:

👍 Nice and simple!

---

> Not clear (to me, anyway) why the test and training images are being treated differently.
> I would call loud attention to the fact -- surprising to newcomers and characteristic of NumPy -- that the division is not truncating the result into uint8's (which would leave zeros almost everywhere) but silently promoting the dtype into uint64, and an explicit cast is unneeded.
> Don't put the note underneath, and don't use "Note" -- include it in the "Normalize the arrays..." text.

Great point on "division -> float64" - that's excellent and so useful!

And, since the shapes are already as needed, the extra `reshape()` step was unnecessary -  I removed it from the code.

image

Here's the new instruction:

> "2. Normalize the arrays by dividing them by 255 (and thus promoting the data type from `uint8` to `float64`) and then assign the train and test image data variables — `x_train` and `x_test` — to `training_images` and `train_labels`, respectively. To make the neural network model train faster in this example, `training_images` contains only 1,000 samples out of 60,000. To learn from the entire sample size, change the `sample` variable to `60000`."

---

> The opening of "Preprocess the data" is wordy and doesn't explain much.

The section is indeed a bit long but it explains a lot of important stuff - data processing usually takes up a lot of time in machine learning.

Since the section was quite long, I made it more concise while keeping the educational stuff to save the readers from extra googling. 

For example, I think it's important to mention the concepts of tensors, various types of floating point formats, rescaling/normalizing data + the why. 

I've also removed the "You have loaded the MNIST dataset and saved it to disk as 4 files: ..." section to avoid repetition.

> "Neural networks can work with inputs that are in a form of tensors (multidimensional arrays) of floating-point type. When preprocessing the data, you should consider the following processes: [vectorization](https://en.wikipedia.org/wiki/Vectorization_%28mathematics%29) and [conversion to a floating-point format](https://en.wikipedia.org/wiki/Floating-point_arithmetic#Floating-point_numbers).
> 
> "Since the MNIST data is already vectorized and the arrays are of `dtype` `uint8`, your next challenge is to convert them to a floating point format, such as `float64` ([double-precision](https://en.wikipedia.org/wiki/Double-precision_floating-point_format)):
> 
> "- _Normalizing_ the image data: a [feature scaling](https://en.wikipedia.org/wiki/Feature_scaling#Application) procedure that can speed up the neural network training process by standardizing the [distribution of your input data](https://arxiv.org/pdf/1502.03167.pdf).
> "- _[One-hot/categorical encoding](https://en.wikipedia.org/wiki/One-hot)_ of the image labels.
> 
> "In practice, you can use different types of floating-point precision depending on your goals and you can find more information about that in the [Nvidia](https://blogs.nvidia.com/blog/2019/11/15/whats-the-difference-between-single-double-multi-and-mixed-precision-computing/) and [Google Cloud](https://cloud.google.com/blog/products/ai-machine-learning/bfloat16-the-secret-to-high-performance-on-cloud-tpus) blog posts."

---

> This and most of the other "Note" sections should be integrated into the main text. Saying "note" isn't necessary; the typography of the note sections is incongruous; and Google style calls this kind of usage a placeholder and discourages it.

I use this style of "Note" to make certain things stand out, especially if some readers are skimming/scanning the text. Too much plain text can be quite hard to follow, in my experience, and we're limited to making text **bold** and _italic_ in plain Markdown. 

For example, Google Cloud docs go further: https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-zonal-cluster:
image

(there's Markdown doc extension for MkDocs that allows you to do similar stuff too, I think)

---

> in this context it's more idiomatic to say "floating point" without the s 

Good point, I agree! 

I also double checked with [this blog post](https://cloud.google.com/blog/products/ai-machine-learning/bfloat16-the-secret-to-high-performance-on-cloud-tpus) and it's clear that the industry uses "floating point" (singular) and "floating-point format".

---

> Encode the labels and assign the values to new variables:

Nice and simple. Thank you!

---

> Examine a few encoded labels:
> ...and compare to the originals:

👍 Excellent, agree.

---

> It's unclear which part of the preceding 120-word description "this type" refers to. Supervised learning means learning from labeled examples. If you want to explain what supervised learning is, say that. My suggestion, though, is to drop the sentence.

That's a good point. I slotted SL into one of the previous sentences ("inputs", "labels" -> SL): _"Based on the image inputs and their labels ([supervised learning](https://en.wikipedia.org/wiki/Supervised_learning)), your neural network will be trained to learn their features using forward propagation and backpropagation..."_

---

> I'd leave backpropagation out of the first line; you say later that it trains using forward propagation and backpropagation. Also typo in "hardwritten."

I agree and thanks for spotting this.

---

> > The final output of the network is a vector of 10 scores — one for each handwritten digit image.
> It's unnecessary and confusing to introduce this detail here. I suggest leaving it out.

I believe this is quite an important detail in the architecture, so I'm going to leave it here. Some way more complex models may be structured with multiple outputs (and/or multiple inputs) e.g. input (image) -> output (description and image) or input (app activity) -> output (various user characteristics).

---

> This tutorial was adapted from the work by Andrew Trask (with the author's permission).

Since I used the code from the GitHub repo, I'd keep the original sentence. I mentioned the book in the suggested reading due to its focus on NumPy.

---

Thanks for the awesome feedback!

content/tutorial-deep-learning-on-mnist.ipynb
8bitmp3: Sounds good @melissawm

8bitmp3: 😄 I think I found the missing bracket.

8bitmp3: OK! I think the new version of this section looks better following the feedback below.

8bitmp3: The epoch number used to be printed every 10th time - it can be less overwhelming as you increase the epochs to thousands. 

I suggest we use your simplified version of the code and write `"Epoch: " + str(j+1)` since we start at `0`, so the count ends at `100` instead of `99`.

I checked the "before" and "after" error and accuracy rates and they appear to be the same — not great (because this is a simple example) but relatively consistent.

Thanks for the revised code! 🙏

content/tutorial-deep-learning-on-mnist.ipynb
8bitmp3: With this one, I moved it to the bottom. That repo helped find an alternative to the `keras` shortcut for dataset downloading, but since then you've revised that part of the code quite a bit 👍

8bitmp3: Got it 😃 1, 2, 3, 4 (not 5)

8bitmp3: That's a good idea. Let me try a side-by-side (left-to-right) layout with `subplots`.

content/tutorial-deep-learning-on-mnist.ipynb
8bitmp3: Fyi @melissawm I'm reusing the main image (a diagram with arrows for forward/back prop + all the details) again in _Model architecture and training summary_ - the section is quite verbose and a picture can be worth a 1000 words 📸

melissawm: Is this 

> @melissawm Can you please double check the code and comments, just in case +1

All good!

melissawm:
> @melissawm Can you please double check the code and comments, just in case +1

All good!

8bitmp3:
fyi @melissawm Changing the small one-hot encoding function to reduce the amount of code (less engineering):

Before:

```python
def one_hot_encoding(labels, dimension=10):
    one_hot_labels = np.zeros((len(labels), dimension))
    for i, label in enumerate(labels):
        one_hot_labels[i][label] = 1 
    return one_hot_labels
```

After:

```python
def one_hot_encoding(labels, dimension=10):
    one_hot_labels = (labels[..., None] == np.arange(dimension)[None])
    return one_hot_labels.astype(np.float64)
```

(I noticed this method when studying [Flax](https://github.com/google/flax) (that's built on JAX, which is NumPy with GPU support 😸))

melissawm:
Hello, folks! Unless there are any blocking issues, I'd like to move forward and merge this tutorial soon. Once no more changes are requested, we can convert it to the new workflow and if any issues remain we can do a follow-up.

melissawm:
Alright folks - I'm merging as is, and will "mistify" it in a subsequent PR. If you have any standing issues please open a new PR after that. Thanks all!

PR #47, opened 2020-11-26
Converts "Deep Learning on MNIST" tutorial to md.

melissawm:
@rossbar can you please take a look when you have the time, let me know if this is the right way to do this?

@8bitmp3 if you see anything wrong with the formatting or content, please let me know.


melissawm:
There is a remaining issue that would be nice to sort out, which is that the last code cell times out when executed via make. Is there a way we can skip that cell when making the HTML site? (We can include a markdown comment about the outputs for users but the cell doesn't have to get executed automatically when building the site). This will surely happen again in the Pong tutorial as well.

rossbar (approved):
@melissawm yup that's exactly it :+1: 

The only issue that cropped up for me was an execution timeout - this example has significantly longer runtimes than the others.

Fortunately, you can control this at the cell, notebook, or project-level with various metadata/configuraiton options. I added a longer timeout to the conf.py in be87354 to demonstrate this. We may need to expand that further when we deploy via CI (which I hope to work on today/tomorrow).

melissawm:
Great, thanks @rossbar !

PR #97, opened 2021-08-12
Add alt-text to image in content/tutorial-style-guide.md

meghanrjones:
This PR adds alternative text to the image with Daniele Procinda's documentation system in content/tutorial-style-guide.md.

### DIFF
--- a/content/tutorial-style-guide.md
+++ b/content/tutorial-style-guide.md
@@ -13,7 +13,7 @@ kernelspec:
 
 # Learn to write a NumPy tutorial
 
-![image](https://documentation.divio.com/_images/overview.png)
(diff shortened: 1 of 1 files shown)


rossbar (commented):
When looking at this I realized that there is a newer link to this material that we should probably be using instead: https://diataxis.fr/. This also gives a name to the documentation system, Diataxis framework, which will make it easier to refer to.

meghanrjones:
> When looking at this I realized that there is a newer link to this material that we should probably be using instead: https://diataxis.fr/. This also gives a name to the documentation system, Diataxis framework, which will make it easier to refer to.

OK, I can update accordingly but will not be able to do so right away.

PR #114, opened 2022-01-04
Fix image display and clean up for NLP tutorial

melissawm:
Adresses https://github.com/numpy/numpy-tutorials/pull/105#issuecomment-1004524990


rossbar (approved):
+1 for switching to markdown syntax for these images.

FWIW, [you can force `myst_parser` to recognize html-formatted images with the `html_image` extension](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#images), but avoiding html is better IMO.

PR #116, opened 2022-01-18
Update Code of Conduct URL

jsdodge:
The original link to the Code of Conduct was bad. This update points to what I believe to be the correct (and stable) URL.


rossbar (approved):
Thanks @jsdodge , this does indeed seem like the correct CoC to point to. Nice catch!

PR #118, opened 2022-01-26
Bump Gym Requirement Version for M1 Support

saulshanabrook:
I had to bump the gym requirement to 0.18.3, so that it required a version of pillow > 8.1.1 which added wheels for m1 (https://github.com/python-pillow/Pillow/issues/5093#issuecomment-788459509).


rossbar (approved):
Thanks @saulshanabrook

PR #124, opened 2022-01-30
MAINT: Remove Matplotlib as a dependency for AQI tutorial

Mukulikaa:
I had made a plot with Matplotlib in an earlier version of the tutorial but didn't end up using it for the final version and I forgot to remove the corresponding lines.


rossbar (approved):
Thanks for the touchups @Mukulikaa !

PR #125, opened 2022-02-02
Move Jupyter Book to a conda forge requirement

saulshanabrook:
Jupyter Book has a conda forge package, so it can be installed from there instead of pip, to help with dependency resolution.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -12,6 +12,6 @@ dependencies:
   - pip
   - imageio
   - pooch
+  - jupyter-book
   - pip:
-    - jupyter-book
     - gym[atari]==0.19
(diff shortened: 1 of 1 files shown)


rossbar (commented):
Actually it turns out that jupyter-book isn't a dependency at all - the dependencies for building the site are listed in `site/requirements.txt`. Would you mind updating the environment.yml to use those instead?

I suspect the `environment.yml` has gotten out-of-date because it isn't used/tested anywhere. We should probably add a CI job for that...

saulshanabrook:
@rossbar, sure I can do that.

saulshanabrook:
@rossbar done!

rossbar (approved):
Thanks @saulshanabrook , LGTM!

PR #167, opened 2023-01-15
Fix link in svd tutorial

OriolAbril:


rossbar (approved):
Thanks @OriolAbril ! I took the liberty of removing the added wording at the end of the sentence for brevity's sake since the original wording still works IMO. The updated link and renamed link name remain - thanks again!

PR #217, opened 2024-06-18
Add Jupytext requirement for binder.

Carreau:
I think binder looks at env.yaml not at requirements.txt maybe that's the reason that binder don't have jupytext installed

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -14,3 +14,5 @@ dependencies:
   - myst-nb
   - sphinx-book-theme
   - sphinx-copybutton
+  # to load the md files in binder
+  - jupytext


melissawm (approved):
Let's try - I think it makes sense. Thanks @Carreau !

melissawm:
Looks like this works - in any case we still need to "Open as notebook"
Captura de imagem_20240618_153108

bsipocz:
what about ` jupyterlab-myst`, would that fix this?

rowanc1:
You will still need `jupytext`, and to open as notebook with `jupyterlab-myst`, the tools are designed to work together. If you want the experience for users to be simpler, you could potentially render out the (pre-executed) notebooks, and have that be the Jupyter experience that gets started.

bsipocz:
> you could potentially render out

you mean an ipynb notebook? (That is what I do with astroML and really don't like the process/setup to have both mb and ipynb (but also, I need the outputs at least for a couple of notebooks as some are too expensive to execute each time in CI)

rowanc1:
Yes, that is what I meant is automatically creating those for when people are interacting with the binder instance. I am not suggesting keeping those in git though! That would require some improvements/modifications to CI to enable that though.

melissawm:
We are doing something similar here: https://github.com/melissawm/scipy/blob/d304e95751d52dd327e10634b06c17f0440ba5bc/doc/Makefile#L116

bsipocz:
We do that here, too, they are then pushed to the `gh-pages` branch along with the rendered htmls. Could binder be forced to pick up those ones?

rossbar:
I'd rather pursue client-side interactivity (ideally on the page itself) than put effort into adding `.ipynb`-formatted files for binder/jupyterhub support. I'd like to push towards a future where `.md`-formatted notebooks are natively supported. Jupytext is fantastic, and the implicit behind-the-scenes support for md->ipynb that makes this work currently is a critical component in a lot of the sustainable docs workflows (not only for server-side interactivity, but also for testing/linting, etc.).

melissawm:
I agree, I think that would be ideal. But as a compromise while that doesn't happen we can have binder point to the ipynb files. I'll try to poke around.

rowanc1:
In the new `mystmd` work we have put a lot of work into `thebe` which enables client-side interactivity (powered by jupyter, binder or jupyterlite). There are some docs on that here:

https://mystmd.org/guide/integrating-jupyter

Because the numpy tutorials are using well-supported packages, I think that adding `jupyter: lite` to the myst project config, might just work..?

We currently haven't enabled editing directly on the page yet, but it is on our roadmap.

- https://github.com/executablebooks/mystmd/issues/443

Curious if @rossbar and @melissawm this is getting close to the direction you imagine?

rossbar:
> Curious if @rossbar and @melissawm this is getting close to the direction you imagine?

Indeed - this was one of the features from that was most compelling from my perspective! It's great to see that on-page interactivity is on the roadmap!

melissawm:
Yes, 100%! Should we follow that issue for updates?

rowanc1:
I am tracking the request from MyST's side here:

- https://github.com/executablebooks/mystmd/issues/1270

bsipocz:
@rossbar @melissawm - FYI: Angus helped me setting up a repo with markdown notebooks to work on binder out of the box, I'm happy to propagate the solution here while we're working out the switch from sphinx and static pages.

PR #248, opened 2025-02-06
Update and rename README.md to README_ES.md

luiscargarcia:


melissawm:
Hi @luiscargarcia , it looks like you are trying to translate the README to Spanish, but with this PR you are erasing the english version of the README which is not correct.

To be clear, we currently don't have a workflow set up to translate these tutorials. If you are interested in translating the NumPy website (excluding the documentation) you can take a look at https://github.com/numpy/numpy.org/issues/55 or more broadly https://scientific-python-translations.github.io/.

Cheers!

PR #260, opened 2025-08-08
Slightly improve explanation of image shape in svd-tutorial.md

### LINKED ISSUE #259: Misleading axis interpretation in SVD tutorial
The [SVD tutorial](https://numpy.org/numpy-tutorials/content/tutorial-svd.html) interprets the shape of the image in the following way:

> img.shape
> 
> (768, 1024, 3)
> 
> The output is a [tuple](https://docs.python.org/dev/tutorial/datastructures.html#tut-tuples) with three elements, which means that this is a three-dimensional array. In fact, since this is a color image, and we have used the `imread` function to read it, the data is organized in three 2D arrays, representing color channels (in this case, red, green and blue - RGB). You can see this by looking at the shape above: it indicates that we have an array of 3 matrices, each having shape 768x1024.

I think it is misleading to explain the organisational structure of the array as consisting of three 2D arrays. The data is not organized as an array of 3 matrices. 

In numpy, the leftmost axis is the outermost. In shape `(768, 1024, 3)` the colour channels axis stands for the innermost axis.
So rather than "three matrices of shape 768x1024," I think it helps building intuition better to talk about  a 768x1024 grid of pixels, where each pixel has 3 values for RBG.

Do others agree with this? If so, I could make a PR that suggests a better wording.

StefanieSenger:
This PR re-words a paragraph on shape and axes of an image.

closes #259

### DIFF
--- a/content/tutorial-svd.md
+++ b/content/tutorial-svd.md
@@ -80,7 +80,7 @@ First, let's check for the shape of the data in our array. Since this image is t
 img.shape
 ```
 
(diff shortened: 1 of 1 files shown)


melissawm (approved):
Thanks!

PR #301, opened 2026-01-16
Rename index.md to index.mdx

MichaelBiegluk:


mattip:
Why would you do this and break the site?

